PR #11, opened 2015-10-30
[CSS] Fixed something for the #views[device=all]

limhenry:
A horizontal scroll bar will appear in the #views on certain screen size. Fixed by using "max-width"


limhenry:
I signed it!

frankiefu:
I would like to retain the margin so it looks better.  We could just set `overflow: hidden` on the canvas, e.g.
```
.canvas[device=all] {
  overflow: hidden;
}
```

PR #12, opened 2015-10-30
Add instructions for cloning carbon-elements

justinfagnani:


bower.json
```diff
@@ -12,7 +12,6 @@
     "iron-resizable-behavior": "polymerelements/iron-resizable-behavior#^1.0.0"
   },
   "devDependencies": {
-    "carbon-elements": "polymerelements/carbon-elements#gh-pages",
```
frankiefu: I think we can just change this to
```
"carbon-elements": "git@github.com:PolymerElements/carbon-elements.git",
```
This seems to work for me given I have ssh key setup on my desktop.

addyosmani: I would recommend switching to Frankie's suggestion. `gh-pages` is the default branch on that repo and just specifying the raw git URL is sufficient to pull it down.

frankiefu:
`lab-router` is now in a public repo (https://github.com/PolymerLabs/lab-elements)

PR #37, opened 2016-01-04
App Drawer

keanulee:


app-drawer/app-drawer.html
```diff
@@ -120,37 +116,156 @@
...
+      _persistentChanged: function() {
+        this._updateDocScroll();
+      },
+
+      _translateDrawer: function(x) {
+        this.transform('translateX(' + x + 'px)', this.$.drawerContent);
+      },
+
+      _updateDocScroll: function() {
+        this.shouldEnableDocScroll(this.persistent || !this.opened);
```
keanulee: `overflow: hidden` on supported browsers (not mobile Safari); remove `AppDocumentScrollBehavior`

app-drawer-layout/app-drawer-layout.html
```diff
@@ -243,53 +149,36 @@
...
+        this._setNarrow(narrow);
+        this._resetLayout();
       },
 
-      _drawerActivate: function() {
-        if (!this.disableDrawerActivate && this.narrow) {
-          this.async(function() {
-            this.opened = false;
-          });
+      _resetLayout: function() {
```
keanulee: Should listen for `<app-drawer>` app-drawer-position-changed event and update marginLeft/Right

app-drawer/app-drawer.html
```diff
@@ -42,75 +44,69 @@
...
-    }
+      -webkit-transition-timing-function: var(--app-drawer-timing-function, ease);
+      transition-timing-function: var(--app-drawer-timing-function, ease);
 
-    :host([position=right]) {
-      right: 0;
-      left: auto;
-      -webkit-transform: translate3d(100%, 0, 0);
-      transform: translate3d(100%, 0, 0);
+      @apply(--app-drawer-styles);
```
keanulee: Rename to `--app-drawer-content`

app-drawer/app-drawer.html
```diff
@@ -42,75 +44,69 @@
...
     }
 
-    :host([position=top]) {
-      top: 0;
+    :host[position=right] #drawerContent {
```
keanulee: Need brackets to work in native ShadowDOM:

```
:host([position=right])
```

Here and everywhere

app-drawer/app-drawer.html
```diff
@@ -42,75 +44,69 @@
...
-      -webkit-transform: translate3d(0, 0, 0);
-      transform:  translate3d(0, 0, 0);
+    :host[opened] #drawerContent {
+      -webkit-transform: translate(0, 0);
+      transform:  translate(0, 0);
+    }
 
-      visibility: visible;
+    #scrim {
+      @apply(--app-scrim-styles);
```
keanulee: `--app-drawer-scrim`

app-drawer/app-drawer.html
```diff
@@ -42,75 +44,69 @@
...
       right: 0;
-      bottom: auto;
-      left: 0;
-      width: auto;
-      height: 256px;
-      -webkit-transform: translate3d(0, -100%, 0);
-      transform: translate3d(0, -100%, 0);
+
+      -webkit-transform: translate(100%, 0);
+      transform: translate(100%, 0);
```
keanulee: `translate3d` for own painting layer (not `will-change` for browser compatibility), here and everywhere.

app-drawer/app-drawer.html
```diff
@@ -120,37 +116,156 @@
...
+       * @return {number} The width of the drawer in pixels.
+       */
+      getDrawerWidth: function() {
+        return this.$.drawerContent.offsetWidth;
+      },
+
+      _tapHandler: function() {
+        if (!this.persistent) {
+          this.opened = false;
+        }
```
keanulee: Don't close on drawer tap - only close when tap on scrim

app-drawer-layout/app-drawer-layout.html
```diff
@@ -178,45 +133,6 @@
           value: '640px'
         },
```
keanulee: Consider `/s/narrow/isNarrow/` and `/s/forceNarrow/narrow/`

app-drawer/app-drawer.html
```diff
@@ -120,37 +116,156 @@
...
       toggle: function() {
         this.opened = !this.opened;
+      },
+
+      /**
+       * Gets the width of the drawer.
+       *
+       * @return {number} The width of the drawer in pixels.
+       */
+      getDrawerWidth: function() {
```
keanulee: `getWidth()`

app-scrim/app-scrim.html
```diff
@@ -44,7 +44,7 @@
       background-color: var(--app-scrim-color, black);
       visibility: hidden;
 
-      transition: 0.5s ease-out;
+      transition: 0.2s ease-out;
```
keanulee: Just delete `<app-scrim>` and transition a `div#scrim` in `<app-drawer>`

app-drawer-layout/app-drawer-layout.html
```diff
@@ -224,17 +139,8 @@
...
-      },
-
-      /**
-       * Toggles the drawer panel open and closed.
-       *
-       * @method toggleDrawer
-       */
-      toggleDrawer: function() {
-        this.opened = !this.opened;
+        this.listen(this.drawer, 'app-drawer-position-changed', '_resetLayout');
```
frankiefu: this assumes the drawer is there at `app-layout-drawer`'s attached time.  It is possible that the drawer is appended dynamically and this will not work.  How about just add the `app-drawer-position-changed` to the `listeners` object?

keanulee: SGTM

app-drawer/app-drawer.html
```diff
@@ -120,39 +145,182 @@
...
 
         /**
-         * Determinates the position of the drawer on the screen.
-         * Options are left, right, top, bottom.
+         * The drawer does not have a scrim and cannot be swiped close.
+         */
+        persistent: {
+          type: Boolean,
+          value: false,
+          reflectToAttribute: true,
```
frankiefu: Should consider adding `reflectToAttribute: true` to property that really really needs it.  It has perf impact on load time.

keanulee: Yes, I was mindful of limiting usage of `reflectToAttribute`. In this particular case, it is used for styling.

demo/demo3.html
```diff
@@ -52,11 +53,11 @@
...
+    <app-drawer>
       <app-toolbar>App name</app-toolbar>
-    </div>
+    </app-drawer>
 
-    <app-header-layout main has-scrolling-region>
+    <app-header-layout has-scrolling-region>
 
       <app-header condenses reveals effects="waterfall">
         <app-toolbar>
```
frankiefu: When I tried to run the demo, the main content shows empty.  Can you check to see if is related to the changes?

keanulee: Ah, tried to remove some CSS from app-drawer-layout but removed too much. Added back support for the `[fullbleed]` attribute and relative positioning/100% height for its content container.

frankiefu:
LGTM :+1:

app-drawer-layout/app-drawer-layout.html
```diff
@@ -156,7 +119,6 @@
           type: Boolean,
           readOnly: true,
           notify: true,
-          reflectToAttribute: true,
```
frankiefu: I think we actually want to keep this one so one can style stuffs based on the existence of `narrow` attribute.  For example, in the publishing template here: https://github.com/PolymerLabs/app-layout/blob/drawer/templates/publishing/src/blog-app.html#L100

keanulee: When I first saw this I thought that that `blog-app` styling would not work because of `:not([attribute])` selectors not matching properly in IE/Edge, but for some weird reason (selector matches a class as well, maybe?) it does work.

I'll add the `reflectToAttribute` back to `narrow`.

PR #77, opened 2016-02-23
Make entire app-drawer have fixed positioning

keanulee:
This allows users to set `z-index` on the `<app-drawer>` element instead of having to set it on both its scrim and content.


frankiefu:
lgtm

frankiefu:
I need to revert this cause it was breaking apps like pesto when in desktop layout.

PR #79, opened 2016-02-23
Fixed persistent drawer that doesn't cover content

keanulee:
Let’s try this again… this should not cover the content when the drawer is persistent

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -41,6 +41,7 @@ content, and apply height and overflow styles to it.
 
 Custom property                  | Description                            | Default
 ---------------------------------|----------------------------------------|--------------------
+`--app-drawer-width`             | Width of the drawer                    | 256px
 `--app-drawer-content-container` | Mixin for the drawer content container | {}
 `--app-drawer-scrim-background`  | Background for the scrim               | rgba(0, 0, 0, 0.5)
 
@@ -52,6 +53,11 @@ Custom property                  | Description                            | Defa
   <style>
 
     :host {
+      position: fixed;
+      top: -120px;
+      bottom: -120px;
+      left: 0;
+      right: 0;
       visibility: hidden;
       transition: visibility 0.2s ease;
     }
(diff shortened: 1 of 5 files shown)


app-drawer/app-drawer.html
```diff
@@ -108,24 +126,41 @@
...
+      position: absolute;
+      top: 0;
+      bottom: 0;
       left: 0;
       right: 0;
-      bottom: -120px;
       background: var(--app-drawer-scrim-background, rgba(0, 0, 0, 0.5));
       opacity: 0;
-      transition: opacity 0.2s ease;
+      visibility: hidden;
```
frankiefu: Do we need to set `visibility: hidden` on #scrim?  Seems like we don't need to do that here and also L159.

keanulee: Yes - if not, the scrim flashes visible when toggling `app-drawer-layout` from split to narrow layout (I broke this in a previous commit). Example demo/demo6.html.

frankiefu: Hmm... I commented out [L136](https://github.com/PolymerLabs/app-layout/blob/fixed-drawer-v2/app-drawer/app-drawer.html#L136) and [L159](https://github.com/PolymerLabs/app-layout/blob/fixed-drawer-v2/app-drawer/app-drawer.html#L159) and then ran demo/demo6.html, I couldn't see the flashes when toggling the layout.

app-header/test/app-header.html
```diff
@@ -48,7 +48,7 @@
   <test-fixture id="testHeader">
     <template>
       <div>
-        <app-header>
+        <app-header class="header">
```
frankiefu: Hmm... this sounds like a bug in Polymer.  Users should be able to override the :host rule.  Let me try to reproduce it in a simple testcase.

frankiefu: https://github.com/Polymer/polymer/issues/3459

frankiefu:
lgtm

PR #86, opened 2016-02-26
update zuperkulblog

frankiefu:
- remove tap highlight color on iOS
- remove :hover on article which seems to make scrolling a bit more jank
- add min-height to page to make it looks a bit better for empty page

### DIFF
--- a/templates/publishing/src/blog-app.html
+++ b/templates/publishing/src/blog-app.html
@@ -40,6 +40,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       :host {
         --app-primary-color: #404040;
+        -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
       }
 
       app-drawer {
@@ -104,15 +105,15 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       app-drawer-layout:not([narrow]) .main-title-toolbar {
         width: 100%;
       }
+      
+      .category-page {
+        min-height: 100vh;
+      }
 
       [is=article-headline] {
         @apply(--shadow-transition);
         @apply(--shadow-elevation-2dp);
         cursor: pointer;
-      }
-
-      [is=article-headline]:hover {
-        @apply(--shadow-elevation-6dp);
       }
 
       article-detail {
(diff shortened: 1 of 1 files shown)


templates/publishing/src/blog-app.html
```diff
@@ -40,6 +40,7 @@
 
       :host {
         --app-primary-color: #404040;
+        -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
```
abdonrd: I propose:

```
/* NOTE: Both values are needed, since some phones require the value to be `transparent`. */
-webkit-tap-highlight-color: rgba(0, 0, 0, 0);
-webkit-tap-highlight-color: transparent;
```

Reference: https://github.com/PolymerElements/paper-fab/pull/48#discussion_r50937454

keanulee: Ref https://github.com/PolymerElements/paper-fab/pull/48#discussion-diff-50608788

```css
        /* NOTE: Both values are needed, since some phones require the value `transparent`. */
        -webkit-tap-highlight-color: rgba(0,0,0,0);
        -webkit-tap-highlight-color: transparent;
```

frankiefu: From the thread I'm still not clear why do we need `-webkit-tap-highlight-color: transparent;`.  iOS recommends to [disable tap highlighting, set the alpha to 0](https://developer.apple.com/library/mac/documentation/AppleApplications/Reference/SafariWebContent/AdjustingtheTextSize/AdjustingtheTextSize.html).  And it seems to me `-webkit-tap-highlight-color: rgba(0, 0, 0, 0)` is enough and works on our supported iOS Safari browsers. 

Is there any browser we support that we need to use `-webkit-tap-highlight-color: transparent`?

keanulee: IIRC `transparent` is required for some Android phones/browsers. If it works on the most recent Chrome though, I think it's okay to remove it.

frankiefu: right... it was for old Chrome on Android when Chrome was using webkit

keanulee:
LGTM

templates/publishing/src/blog-app.html
abdonrd: Then we can remove the `-webkit-tap-highlight-color: transparent` on `paper-elements`?

keanulee: I think app-layout is in a more experimental stage, so it's safe to remove it here. I would leave it in the paper-elements for now since it's not too burdensome right now.

abdonrd: :+1:

PR #96, opened 2016-03-02
Fix drawer calculate velocity method

keanulee:
Needed for Chrome 49 - see https://developers.google.com/web/updates/2016/01/high-res-timestamps?hl=en

Should probably be added in IOWA ASAP, so probably should do a release after merging (if there no other breaking changes on master)

cc @frankie @blasten


frankiefu:
lgtm

PR #107, opened 2016-03-09
Refocus to app-drawer itself and close with esc

keanulee:


app-drawer/app-drawer.html
```diff
@@ -170,6 +170,10 @@
 
       is: 'app-drawer',
 
+      hostAttributes: {
+        tabindex: '0'
+      },
+
```
frankiefu: What about setting `tabindex="0"` on the `#contentContainer`?  If this is something we don't really want to expose to the user then it's better to not put it on the host.

keanulee: Sure - I don't see a problem with that.

app-drawer/app-drawer.html
```diff
@@ -318,19 +322,19 @@
       },
 
       _focusHandler: function(event) {
-        if (!this.isLightDescendant(Polymer.dom(event).rootTarget)) {
-          this._focusInsideDrawer();
+        if (Polymer.dom(event).path.indexOf(this.$.contentContainer) === -1) {
```
frankiefu: :+1:

keanulee:
@frankiefu A user can listen for transitionend on `<app-drawer>` itself and focus on that. There are a total of 3 transitionend events (app-drawer visibility, scrim opacity, and contentContainer transform), but that can be easily read by the user.

frankiefu:
`transitionend ` is implementation detail and I would prefer to have an event clearly indicate when the drawer is completed closed/opened.

frankiefu:
Let's think about if we want to add a new event later.  LGTM on this PR.

Issue #121, opened 2016-03-18
app-drawer - position-fixed limits drop in use cases


ktiedt:
- Allow for position other than fixed
- Animate width (maybe?) - to allow parent node to collapse around it.

What this allows for is 2 fold.
1. app-drawer can now be dropped into an existing nav container and expand/collapse that container
2. becomes a drop in replacement for non floating/blocking menus.

ktiedt:
I think I have an alternative solution for this besides causing layout events due to the width changing. Should have a PR later today if it all works out.

keanulee:
@ktiedt Did #132 help with this, or do you need more changes?

ktiedt:
We are good, thank you!

PR #124, opened 2016-03-21
[DO NOT MERGE] Added a few events to app-drawer

ktiedt:
app-drawer-opened/closed - Fires immediately upon opened state toggling.
app-drawer-toggled - Fires immediately + {opened: this.opened} evt.detail property.

app-drawer-transitioned-opened/closed - Fires after the CSS animation for the drawer completes.

This allows older UIs that are updating to Polymer a means to tie into the app-drawer's state to affect their non Polymer UI as well, both with and without animations being taken into account.

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -179,7 +179,8 @@ Custom property                  | Description                            | Defa
           type: Boolean,
           value: false,
           notify: true,
-          reflectToAttribute: true
+          reflectToAttribute: true,
+          observer: '_openedChanged',
         },
 
         /**
@@ -255,10 +256,12 @@ Custom property                  | Description                            | Defa
         // Only transition the drawer shortly after it is attached (e.g. app-drawer-layout
         // may need to set the initial opened state which should not be transitioned).
         requestAnimationFrame(this._setTransitionDuration.bind(this, ''));
+        this._setupTransitionEnd(false);
       },
 
       detached: function() {
         document.removeEventListener('keydown', this._boundKeydownHandler);
+        this._setupTransitionEnd(true);
       },
 
       /**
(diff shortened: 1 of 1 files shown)


ktiedt:
These changes in conjunction with wrapping app-drawer in a secondary custom-element (or using custom-styles) allows us to resolve all the issues in #121

app-drawer/app-drawer.html
```diff
@@ -535,6 +538,39 @@
         this.fire('app-drawer-overlay-changed', { overlay: this.opened && !this.persistent });
       },
 
+      _openedChanged: function(opened) {
+        var evtName = this.opened ? 'opened' : 'closed';
+
+        this.fire('app-drawer-' + evtName);
```
keanulee: Instead of making new events, can you rely on app-drawer's `opened-changed` event, then read the app-drawer's `opened` property (`opened-changed` is provided since there's `notify: true`, see https://www.polymer-project.org/1.0/docs/devguide/data-binding.html#change-notification-protocol)

app-drawer/app-drawer.html
```diff
@@ -535,6 +538,39 @@
...
+      },
+
+      _onTransitionEnd: function(evt) {
+        var opened = this.opened;
+        var evtName = this.opened ? 'opened' : 'closed';
+
+        this.fire('app-drawer-transitioned-' + evtName);
```
keanulee: Similarly, rely on `app-drawer-overlay-changed` (which handles de-duplicating `transitionend` events, since more than one child is transitioned - see https://github.com/PolymerLabs/app-layout/blob/master/app-drawer/app-drawer.html#L556), then read the `opened` property

app-drawer/app-drawer.html
ktiedt: That actually would probably work, not sure why I didn't think of this actually, been a long day!

app-drawer/app-drawer.html
ktiedt: Is that what this is for? The name and documentation did not really give me that impression, nor does the code really.

keanulee: The `app-drawer-overlay-changed` event only fires when the scrim finishes transitioning, and the scrim only transitions when the drawer is an overlay (i.e. it is above and partially obscuring the main content). It notably does not fire if the drawer is `persistent`.

If you are not using `persistent`, then this event will probably work for you. If you are, then we might need something else, but I don't want to cross that bridge unless we need to.

ktiedt: persistent is a must in our case. Which drove my current code.

keanulee: I see - I think the best thing to do would be to update the code for the `app-drawer-overlay-changed` event such that it relies on contentContainer instead of the scrim. We need to be careful though since we only want to fire the event once - there is no transitionend if you drag the drawer all the way open then let go, but there is when you drag it all the way closed since `translateX(-100%)` != `translateX(-256px)`. That's why there's:

```javascript
        // Trigger drawer overlay changed now since there will be no transitionend event.
        if (isInEndState) {
          this._drawerOverlayChanged();
        }
```

which is fine right now since it's the scrim, but that will have to change if it's for contentContainer.

I'll take a look at implementing this myself, but I just want clarification that this is what you need. Thanks.

keanulee:
Please do not merge this - see #132

keanulee:
Closing now that #132 has been merged.

PR #132, opened 2016-03-22
Replace app-drawer-overlay-changed event with app-drawer-transitioned

keanulee:
Should cover the use cases in #124

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -539,7 +539,7 @@ Custom property                  | Description                            | Defa
         }
 
         if (oldState !== this._drawerState) {
-          if (this.opened && !this.persistent) {
+          if (this._drawerState === this._DRAWER_STATE.OPENED) {
             document.addEventListener('keydown', this._boundKeydownHandler);
             document.body.style.overflow = 'hidden';
           } else {
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -157,7 +157,7 @@ Custom property                  | Description                            | Defa
 
   <template>
 
-    <div id="scrim" on-tap="close" on-transitionend="_scrimTransitionend"></div>
+    <div id="scrim" on-tap="close"></div>
 
     <div id="contentContainer">
       <content></content>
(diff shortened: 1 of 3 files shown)


keanulee:
@ktiedt Will something like this work for you?

ktiedt:
Just tested, and this does get us to a usable state. In actually the on-opened-changed pointer was all we needed in the end (decided to not try to match the animation effects since we don't use them anywhere else yet).

I would suggest adding an object with the state to the transition event if it is to be the only transition event though. That way a user does not have to dig through e.currentTarget to access things that should be relevant to the event itself.

Thanks!

keanulee:
@ktiedt What details do you think would be helpful? I didn't want to expose the private `_DRAWER_STATE` object, but perhaps including the current state of opened and persistent in e.detail (even though they would be accessible through e.target)?

ktiedt:
I generally expect custom-events to contain details relevant to the event, in this case the event w/o details does not answer the question of 'transitioned to what?' - If it were me (as you can see in my PR) I would add the state ie: `{opened: this.opened}` but if you are going with minimalist (which is technically true to what normal HTML events do) approach, making the user dig through the element for this would be on par. In the end either way is usable.

keanulee:
Good point. I think we'll stick with the minimalist approach and leave out the detail. Thanks.

app-drawer/app-drawer.html
```diff
@@ -524,25 +529,47 @@
...
+            this._DRAWER_STATE.OPENED_PERSISTENT : this._DRAWER_STATE.OPENED;
         } else {
-          document.removeEventListener('keydown', this._boundKeydownHandler);
-          document.body.style.overflow = '';
+          this._drawerState = this._DRAWER_STATE.CLOSED;
         }
 
-        this.fire('app-drawer-overlay-changed', { overlay: this.opened && !this.persistent });
+        if (oldState !== this._drawerState) {
+          if (this.opened && !this.persistent) {
```
frankiefu: Is this the same as `if (this._drawerState === this._DRAWER_STATE.OPENED) {`?

keanulee: I guess it is - will change.

frankiefu:
LGTM

ktiedt:
Thanks!

PR #150, opened 2016-04-05
Add IronResizeableBehavior to app-drawer-layout

### LINKED ISSUE #144: app-drawer-layout content not resizing in IE
Hi,
When using an app-drawer inside an app-drawer-layout and we resize the page to make app-drawer disappear, the app-header is not resizing well. 
Step to reproduce:
1)	Go in demo5.html in IE (I used IE 11 for my test)
2)	Click on explorer box to reduce windows size (The bug only appear in a one shot resize, not when resizing windows by the corner).  The internet explorer window should be at a point that make the drawer disappear after clicking the resize button.
3)	We can see that the app-header doesn’t fill the remaining space of the drawer.

Screen shot: 

 
image

Note: I do not only see this bug in app-header. In my app, I have an element that use iron-resize to take the width of the app-header-layout inside the app-drawer-layout and I got the same problem. It seems that the iron-resize event is trigger before the app-drawer have disappear.

keanulee:
Fixes #144

### DIFF
--- a/app-drawer-layout/app-drawer-layout.html
+++ b/app-drawer-layout/app-drawer-layout.html
@@ -10,6 +10,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 <link rel="import" href="../../polymer/polymer.html">
 <link rel="import" href="../../iron-media-query/iron-media-query.html">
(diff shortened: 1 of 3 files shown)


frankiefu:
LGTM

PR #164, opened 2016-04-12
Use IDs in publising/pesto URLs; small fixes

### LINKED ISSUE #163: Update Pesto to use numeric IDs in URLs
...so that it doesn't have special characters (like space) that need to be encoded.

keanulee:
Fixes #163


frankiefu:
LGTM

PR #178, opened 2016-04-19
Perf enhancements for app-drawer/-layout

keanulee:


frankiefu:
LGTM

ebidel:
What are the perf improvements? Is it something IOWA should bring in with a new release? We've frozen to 0.0.15 btw.

frankiefu:
improvement on create/render time in app-drawer & app-drawer-layout, and is about 22% faster.  There is an [issue](https://github.com/PolymerLabs/app-layout/issues/179) in #master and probably do a release after that is fixed.

PR #182, opened 2016-04-20
Focus trap for app-drawer

### LINKED ISSUE #165: Increase drawer swipe sensitivity.
App-drawer swipe to open is not as responsive as paper elements counterpart. On the example demos provided, trying to rapidly swipe on mobile doesn't work or using thumb with slight touch doesn't open the drawer either. 
I'm sure it should be a quick issue to improve.

keanulee:
Fixes #165 as well

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -115,7 +115,7 @@ Custom property                  | Description                            | Defa
 
         visibility: visible;
 
-        width: 15px;
+        width: 20px;
 
         content: '';
       }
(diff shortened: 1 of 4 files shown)


app-drawer/app-drawer.html
```diff
@@ -550,6 +567,66 @@
         }
       },
 
+      _addKeyboardFocusTrap: function() {
+        if (this.noFocusTrap) {
+          return;
+        }
+
+        this._lastFocusedElement = document.activeElement;
```
frankiefu: Under native Shadow DOM, `document.activeElement` returns the outermost element.  e.g. focus anything inside publishing app `document.activeElement` will always return `blog-app`.  Fortunately we can still get the information from shadow root (https://www.w3.org/TR/2015/WD-shadow-dom-20151215/#active-element).  We may need to do something similar like this: https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-manager.html#L68

frankiefu:
LGTM

PR #208, opened 2016-05-23
App-drawer doesn't close with persistent to true

### LINKED ISSUE #207: [app-drawer] shouldn't close with persistent to true
### Description
The `app-drawer` doesn't have to be closed when persistent is true, becuase it's means that the `drawer-layout` is narrow. The old `paper-drawer-panel`, detects if the drawer-panel is narrow and then never close he drawer.
Yo have to check the persistent status every time. Like here: https://github.com/PolymerElements/app-layout/blob/master/templates/publishing/src/blog-app.html#L323-L325

I guess that close method should do this check. If you change the `opened `property then it's your business check persistent property, but calling the method is component responsibility does't close it self.

### Expected outcome

Close function doesn't close the drawer when the menu is persistent.

### Actual outcome

Changing the property closed to true it close itself even with persistent property true.

### Live Demo

https://polymerelements.github.io/app-layout/templates/getting-started/

### Steps to reproduce

In the Live Demo:
1. Open the console
2. Select `app-drawer` node
3. Be sure that property `persistent` is true. (width dependent)
2. Call method `$0.close()`.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari 9
- [x] Safari 8
- [x] Safari 7
- [x] Edge
- [x] IE 11
- [x] IE 10

jorgecasar:
fix #207

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -269,14 +269,20 @@ Custom property                  | Description                            | Defa
        * Closes the drawer.
        */
       close: function() {
-        this.opened = false;
+        if (!this.persistent) {
+          this.opened = false;
+        }
       },
 
       /**
        * Toggles the drawer open and close.
        */
       toggle: function() {
-        this.opened = !this.opened;
+        if (this.opened) {
+          this.close();
+        } else {
+          this.open();
+        }
       },
 
       /**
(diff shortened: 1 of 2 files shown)


keanulee:
See https://github.com/PolymerElements/app-layout/issues/207#issuecomment-221037297

PR #210, opened 2016-05-24
app-drawer 'persistent' attribute should notify

### LINKED ISSUE #209: Persistent attribute should notify
For example, I want to create a drawer button which appears when the drawer is not persistent. Using the opened attribute means the button disappears when the drawer is toggled (causes layout).

I'd like to be able to use persistent={{drawerPersistent}}

samuelli:
Fixes #209

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -192,6 +192,7 @@ Custom property                  | Description                            | Defa
         persistent: {
           type: Boolean,
           value: false,
+          notify: true,
           reflectToAttribute: true
         },
 


keanulee:
See https://github.com/PolymerElements/app-layout/issues/209#issuecomment-221339129

keanulee:
We added a notifying `narrow` property in `app-drawer-layout` in #217 if you want to detect when that changes. Otherwise, the `persistent` property in `app-drawer` would only be changed by the developer, so there shouldn't be a need for it to notify.

PR #217, opened 2016-05-25
Create notifying narrow property on app-drawer-layout

keanulee:
In response to #209

### DIFF
--- a/app-drawer-layout/app-drawer-layout.html
+++ b/app-drawer-layout/app-drawer-layout.html
@@ -143,14 +143,18 @@ Custom property                          | Description                          
         },
 
         /**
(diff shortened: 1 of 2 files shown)


app-drawer-layout/app-drawer-layout.html
```diff
@@ -151,7 +151,11 @@
           value: '640px'
         },
 
-        _narrow: Boolean
+        narrow: {
```
frankiefu: Add doc, e.g. `Returns true if it is in narrow layout. This is useful if you need to show/hide elements based on the layout.`

frankiefu:
minor comment, other LGTM

frankiefu:
LGTM

PR #223, opened 2016-05-30
Removed bottom offset to fix scrolling drawer contents

F4IF:
This fixes the height for the drawer so scrolling contents are handled correctly.  Being new to Polymer I'm not sure if there's a reason for bottom having been -120px (like top), but in my project changing it to 0 didn't have any negative impact.  If there's a reason this must be -120px I propose a variable which can be set to adjust for this.  The only other alternative I had was to hard code a margin-bottom of 120px when I use app-drawer, which doesn't seem very stable.

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -58,7 +58,7 @@ Custom property                  | Description                            | Defa
         position: fixed;
         top: -120px;
         right: 0;
-        bottom: -120px;
+        bottom: 0;
         left: 0;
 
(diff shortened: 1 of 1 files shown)


F4IF:
Signed corporate CLA for Foundation For an Innovative Future (F4IF).

F4IF:
Found some visual artifacts with this change (scroll and background color don't appear to line up).  The correct solution is probably more complicated, so I'm going to close this request.

PR #240, opened 2016-06-13
setScrollDirection after listening for track

keanulee:
Because installing the track listener sets `touch-action: none`


frankiefu:
LGTM

PR #248, opened 2016-06-20
Slight optimizations

keanulee:
Some small things I saw while trying to make `app-drawer-layout` with raw custom elements.


frankiefu:
LGTM

PR #258, opened 2016-07-01
Eliminate duplicate execution of _resetLayout in app-drawer-layout

keanulee:
Before the debounced _resetLayout code in app-drawer-layout would be run twice (!) - this PR should fix it (tested by using console.log statements). Let’s talk about it before merging though, since I’ve tried to fix this a few times before and want to go over this to make sure it’s right

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -211,7 +211,6 @@ Custom property                  | Description                            | Defa
         position: {
           type: String,
           readOnly: true,
-          value: 'left',
           reflectToAttribute: true
         },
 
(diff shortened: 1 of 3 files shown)


app-drawer/app-drawer.html
```diff
@@ -314,9 +313,7 @@
        * @method resetLayout
        */
       resetLayout: function() {
-        this.debounce('_resetLayout', function() {
-          this.fire('app-drawer-reset-layout');
-        }, 1);
+        this.fire('app-drawer-reset-layout');
```
frankiefu: This seems good since now both `app-header` and `app-drawer` fires the `*-reset-layout` event sync: https://github.com/PolymerElements/app-layout/blob/master/app-header/app-header.html#L485

PR #288, opened 2016-08-01
Make app-drawer-layout reset layout on iron-resize

### LINKED ISSUE #283: app-drawer-layout broken content position when hidden during element initialization.
### Description
The `app-drawer` element inside `app-drawer-layout` sits on top of content if hidden when created.  This is due to width being 0 when hidden and resetLayout being called when the element is created instead of when it is shown.

### Live Demo
https://jsbin.com/metolohivi/edit?output

### Steps to reproduce
Open the jsbin demo with a wide browser window (more than 640px).  Note that you cannot see "main title" or "main content".  Resize your window small enough to hide the app-draw, then resize large again.  This manual resize causes 'resetLayout' to run and correct the display issue.  Tested on Chrome 51.

keanulee:
Fixes #283

### DIFF
--- a/app-drawer-layout/app-drawer-layout.html
+++ b/app-drawer-layout/app-drawer-layout.html
@@ -183,9 +183,12 @@ Custom property                          | Description                          
         }
       },
 
+      _notifyingResize: false,
+
       listeners: {
         'tap': '_tapHandler',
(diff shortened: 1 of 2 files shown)


frankiefu:
LGTM

PR #299, opened 2016-08-10
Add opened-when-narrow property to app-drawer-layout

### LINKED ISSUE #278: <app-drawer-layout> overrides <app-drawer> initial values
### Description
`app-drawer-layout` is overriding the initial values for `app-drawer`. 

### Expected outcome
With a narrow screen, if the initial value of `opened` in `app-drawer` is set to `true` the drawer should be showing.

### Actual outcome

With a narrow screen, if the initial value of `opened` in `app-drawer` is set to `true` the drawer is not showing and `opened` is set to false.
### Live Demo

https://jsfiddle.net/rckqb1q7/6/

### Steps to reproduce
1. Have a narrow screen
2. Bind an opened value to `app-drawer`
3. Set the default value to `true`

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari 9
- [x] Safari 8
- [x] Safari 7
- [x] Edge
- [x] IE 11
- [x] IE 10

keanulee:
Fixes #278

### DIFF
--- a/app-drawer-layout/app-drawer-layout.html
+++ b/app-drawer-layout/app-drawer-layout.html
@@ -180,6 +180,14 @@ Custom property                          | Description                          
           type: Boolean,
           readOnly: true,
           notify: true
+        },
+
+        /**
+         * If true, the drawer will initially be opened when in narrow layout mode.
+         */
+        openedWhenNarrow: {
+          type: Boolean,
+          value: false
         }
       },
 
@@ -217,7 +225,8 @@ Custom property                          | Description                          
           var contentContainer = this.$.contentContainer;
 
           if (this.narrow) {
-            drawer.opened = drawer.persistent = false;
+            drawer.opened = this.openedWhenNarrow;
+            drawer.persistent = false;
             contentContainer.classList.add('narrow');
 
             contentContainer.style.marginLeft = '';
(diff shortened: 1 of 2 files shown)


app-drawer-layout/app-drawer-layout.html
```diff
@@ -217,7 +225,8 @@
           var contentContainer = this.$.contentContainer;
 
           if (this.narrow) {
-            drawer.opened = drawer.persistent = false;
+            drawer.opened = this.openedWhenNarrow;
+            drawer.persistent = false;
```
frankiefu: There is an issue with the new change.  For example if you have `openedWhenNarrow` set to true on `app-drawer-layout`, and you run the app on Chrome on Android.  Initially you will see the drawer is opened, which is expected.  Now close the drawer and scroll the document content.  You will see the drawer suddenly pop open.  This is because when document is scrolled, the Chrome address bar will open/close and that triggers a window resize event, which causes the `iron-resize` to fire and lead to `resetLayout` to be called.

Currently we have a few entry points to have `resetLayout` to be called.  I think we should move the code that changes the drawer's state to another function and only call that when `narrow` changes.  For example, this code 
```
drawer.opened = this.openedWhenNarrow;
drawer.persistent = false;
contentContainer.classList.add('narrow');
```
should be moved to another function (e.g. narrowChanged) so we don't have to call this when `app-drawer-layout` receives `iron-resize` event.

frankiefu:
LGTM

PR #313, opened 2016-08-30
Handle cases when app-drawer is appended later

### LINKED ISSUE #273: Uncaught TypeError in polymer-mini.html:2044
### Description

I have the basic app drawer layout setup but I keep getting this from the console.

image

This is the html I have:
```
<app-drawer-layout>
    <app-drawer>
        <app-toolbar style="background-color: #4285f4; color: white;">My App</app-toolbar>
    </app-drawer>

    <app-header-layout>
        <app-header reveals effects="waterfall">
            <app-toolbar>
                <paper-icon-button icon="menu" drawer-toggle></paper-icon-button>
                <div title>{{ title }}</div>
                <paper-icon-button icon="search"></paper-icon-button>
            </app-toolbar>
        </app-header>

        <!-- Main content -->
        <div class="layout vertical" style="padding: 16px;">
            <router-outlet></router-outlet>
        </div>
    </app-header-layout>

</app-drawer-layout>

```

I used https://github.com/vaadin/angular2-polymer to load the component in my angular2 project.

### Expected outcome

I expect no error from the console.

### Actual outcome

The app drawer and everything are functional but I get this error from the console.
polymer-mini.html:2044 Uncaught TypeError: Cannot set property 'persistent' of undefined

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

keanulee:
Might fix #273

### DIFF
--- a/app-drawer-layout/app-drawer-layout.html
+++ b/app-drawer-layout/app-drawer-layout.html
@@ -193,7 +193,8 @@ Custom property                          | Description                          
 
       listeners: {
         'tap': '_tapHandler',
-        'app-drawer-reset-layout': '_handleDrawerResetLayout',
+        'app-drawer-attached': '_resetDrawerState',
+        'app-drawer-reset-layout': 'resetLayout',
         'iron-resize': 'resetLayout'
       },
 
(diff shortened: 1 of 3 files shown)


keanulee:
Actually, this will cause more work if app-drawer is already attached during init - need to revisit

frankiefu:
We might want to handle for the case when app-drawer is upgraded (e.g. lazy import) later.

keanulee:
@frankiefu Does lazy import mean the custom `attached` callback is run after the element is customized? If so, then it should Just Work<sup>TM</sup>.

frankiefu:
Yeah.  Just tried out one of the app-drawer-layout demo and noticed couple things:
- `this.drawer` may return `<app-drawer>` but hasn't been upgraded yet, so at load time it tries to call `this.drawer.getWidth()` and that generates an error.
- because of the error above, `resetLayout` didn't setup the layout correctly.

frankiefu:
LGTM

milky2028:
Hello! I'm currently having this issue using Polymer 2, though it says here it's been resolved. I updated my dependencies, but the issue persists.

milky2028:
This is resulting in the drawer not resetting, so things like 
// Close a non-persistent drawer when the page & route are changed.
        if (!this.$.drawer.persistent) {
          this.$.drawer.close();
        } 
don't work properly. The drawer stays open on page changes.

PR #333, opened 2016-09-20
app-drawer/-layout: Use `click` instead of tap; assignedNodes

keanulee:
`assignedNodes()` doesn't seem to work with Shady DOM. This is causing `app-drawer-layout` tests to fail, even in Canary

### DIFF
--- a/app-drawer-layout/app-drawer-layout.html
+++ b/app-drawer-layout/app-drawer-layout.html
@@ -65,7 +65,7 @@ With an app-header-layout:
 </app-drawer-layout>
 ```
 
-Add the `drawer-toggle` attribute to elements inside `app-drawer-layout` that toggle the drawer on tap events:
+Add the `drawer-toggle` attribute to elements inside `app-drawer-layout` that toggle the drawer on click events:
 
 ```html
 <app-drawer-layout>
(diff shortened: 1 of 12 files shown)


app-drawer-layout/app-drawer-layout.html
```diff
@@ -207,10 +207,12 @@
        * @property drawer
        */
       get drawer() {
-        return Polymer.dom(this).querySelector('[slot=drawer]');
+        return Polymer.Element ?
+            Polymer.dom(this.$.drawerSlot).assignedNodes({ flatten: true })[0] :
+            Polymer.dom(this.$.drawerContent).getDistributedNodes()[0];
```
frankiefu: Yeah, it doesn't work in Shady DOM... let's do something similar to this for now: https://github.com/PolymerElements/iron-dropdown/blob/2.0-preview/iron-dropdown.html#L350

frankiefu:
I ran the app-drawer-layout/demo and `drawer-toggle` doesn't seem to work in FF.

frankiefu:
Also we should update test/index.html to user `?wc-shadydom=true`

keanulee:
@frankiefu 

Clicking `drawer-toggle` doesn't work due to some bug with Shady DOM (`Polymer.dom(event).localTarget` returns `app-drawer-layout, not the button).

Tests have already been updated.

frankiefu:
LGTM

jsilvermist:
Shouldn't `Polymer.dom(event).localTarget` be changed to `event.target` as stated on https://github.com/polymer/polymer/tree/2.0-preview under scoped events?

PR #370, opened 2016-10-24
Don't use this._warn

### LINKED ISSUE #369: Scroll effects using protected `_warn` method not compatible with Closure compiler
Some scroll effects are using the protected `this._warn` method defined by Polymer, which causes Closure warnings. We should use the native `console.warn` method instead

Example:
https://github.com/PolymerElements/app-layout/blob/master/app-scroll-effects/effects/resize-title.html#L32

cc @frankiefu @blasten

keanulee:
Fixes #369 

cc @frankiefu @blasten


frankiefu:
LGTM

PR #371, opened 2016-10-25
Update bower.json

zekida:
"main" field in bower.json should list all the files needed to load this element(like Polymer's bower.json)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,19 @@
     "type": "git",
     "url": "git://github.com/PolymerElements/app-layout"
   },
-  "main": "app-layout.html",
+  "main": [
+    "app-layout.html",
+    "helpers/helpers.html",
+    "app-drawer/app-drawer.html",
+    "app-drawer-layout/app-drawer-layout.html",
+    "app-grid/app-grid-style.html",
+    "app-header/app-header.html",
+    "app-header-layout/app-header-layout.html",
+    "app-scroll-effects/app-scroll-effects-behavior.html",
+    "app-scrollpos-control/app-scrollpos-control.html",
+    "app-toolbar/app-toolbar.html",
+    "app-box/app-box.html"
+  ],
   "dependencies": {
     "iron-flex-layout": "PolymerElements/iron-flex-layout#^1.0.0",
     "iron-media-query": "PolymerElements/iron-media-query#^1.0.0",


zekida:
I signed it!

frankiefu:
[app-layout.html](https://github.com/PolymerElements/app-layout/blob/master/app-layout.html) already includes all the main elements so we don't need to add the files separately.

zekida:
yes we do... that's why i specifically referred [Polymer's bower.json](https://github.com/Polymer/polymer/blob/master/bower.json) as an example.

bower does not parse through imports in 'app-layout.html'. Thats for the html renderer to handle.
the usual bower based frontend tooling pulls the 'main' field as required assets and pushes them to a development server or to a CDN if deploying.

frankiefu:
@zekida Thanks for the detailed explanations.  Now we understand more about the issue and what this PR is trying to solve.  I reopened it and we will take a closer look.  So far it seems pretty reasonable to me.

p.s. In the future it would be good to file an issue so we could have a clear understanding of the issue: https://github.com/PolymerElements/app-layout/blob/master/CONTRIBUTING.md#submitting-pull-requests

zekida:
@frankiefu Thank you for the quick reply, and the reconsideration!
apologies for the abrupt PR, next time I will open an issue beforehand :)

abdonrd:
The paper-input already do that: [PolymerElements/paper-input/bower.json#L13-L21](https://github.com/PolymerElements/paper-input/blob/master/bower.json#L13-L21)

bower.json
```diff
@@ -11,7 +11,19 @@
...
-  "main": "app-layout.html",
+  "main": [
+    "app-layout.html",
+    "helpers/helpers.html",
+    "app-drawer/app-drawer.html",
+    "app-drawer-layout/app-drawer-layout.html",
+    "app-grid/app-grid-style.html",
+    "app-header/app-header.html",
+    "app-header-layout/app-header-layout.html",
+    "app-scroll-effects/app-scroll-effects-behavior.html",
```
keanulee: To be complete, each of the effects in app-scroll-effects/effects/, as well as app-scroll-effects.html, should be listed here

zekida: agreed. sorry for the incomplete PR! thanks!

PR #397, opened 2017-01-17
Add z-index to #contentContainer

limonte:
Currently if there's for example `<google-youtube>` component inside and user tries to maximize that iframe, `<app-drawer>` will overlay maximized iframe:

Live example: https://vaadin.github.io/elements-codelab/#0

This PR will fix that.

### DIFF
--- a/app-drawer-layout/app-drawer-layout.html
+++ b/app-drawer-layout/app-drawer-layout.html
@@ -127,6 +127,8 @@ Custom property                          | Description                          
 
         height: 100%;
 
+        z-index: 1;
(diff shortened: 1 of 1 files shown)


limonte:
ping :)

keanulee:
This won't work - the point of ordering `app-drawer` after the content is so that on narrow viewports (e.g. mobile), the drawer, when opened, appears above the content. Adding `z-index` on the content means that it will appear above instead.

limonte:
Yeah, I see.

PR #401, opened 2017-01-25
2.0-preview auto-generated changes

FredKSchott:


notwaldorf:
test failure looks legit, so this is ok to merge. cc @keanulee as an fyi for the broken test

keanulee:
Only tests failing are on Safari 9 which seem to be passing locally; will continue to monitor these.

PR #411, opened 2017-02-18
fix afterNextRender polymer2 change

nachoab:
Polymer 2 docs:

- Polymer.RenderStatus.afterNextRender(context, callback, args) is now called only with a callback argument (Polymer.RenderStatus.afterNextRender(callback)). The callback should be bound as needed with the correct context and arguments.

https://github.com/Polymer/polymer/blob/2.0-preview/src/utils/render-status.html

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -262,7 +262,7 @@ Custom property                  | Description                            | Defa
         // Only transition the drawer after its first render (e.g. app-drawer-layout
         // may need to set the initial opened state which should not be transitioned).
         this._styleTransitionDuration(0);
-        Polymer.RenderStatus.afterNextRender(this, function() {
+        Polymer.RenderStatus.afterNextRender(() => {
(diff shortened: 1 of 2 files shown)


app-drawer/app-drawer.html
```diff
@@ -262,7 +262,7 @@
         // Only transition the drawer after its first render (e.g. app-drawer-layout
         // may need to set the initial opened state which should not be transitioned).
         this._styleTransitionDuration(0);
-        Polymer.RenderStatus.afterNextRender(this, function() {
+        Polymer.RenderStatus.afterNextRender(() => {
```
TomK: for consistency with the rest of the package it might be preferable to use ES5 function declaration here instead of arrow function

frankiefu:
Thanks for the PR.  The actual fix should be in Polymer and this [PR](https://github.com/Polymer/polymer/pull/4323) has addressed it already.

PR #435, opened 2017-03-18
fix(template): prevent drawer from being swept open on larger screens

### LINKED ISSUE #434: Drawer in Shrine template can always be swept open
### Description
The drawer on the Shrine Template can be swept open at all times, even when it is not supposed to. The drawer should only be able to be swept open on screens <= 400px.

### Expected outcome

The drawer cannot be swept open on larger screens (> 400px).

### Actual outcome

The drawer can always be swept open.

### Live Demo
https://polymerelements.github.io/app-layout/templates/shrine/#latest

GfyCat showing the problem in motion: https://gfycat.com/FloweryMadeupIndianrockpython

### Steps to reproduce

1. Open the [Shrine template](https://polymerelements.github.io/app-layout/templates/shrine/#latest)
2. Ensure your viewport is more than 400px wide
2. Move your cursor to the left edge of the page
4. Click and drag to the right

### Browsers Affected
Tested Chrome, Firefox, and Edge. Presumably happens in all browsers.

Lange:
Fixes #434

This fixes #434 by binding the `swipe-open` attribute of `app-drawer` to the `smallScreen` media query. As a result, `swipe-open` is only enabled when the viewport is <= 400px wide.

I see that the [contributing guidelines](https://github.com/PolymerElements/app-layout/blob/master/CONTRIBUTING.md) state that tests must be added for bugfixes, but it seems that these templates do not have tests.

### DIFF
--- a/templates/shrine/src/shrine-app.html
+++ b/templates/shrine/src/shrine-app.html
@@ -216,7 +216,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     <app-drawer-layout drawer-width="288px" force-narrow>
       <!-- navigation drawer for small screen sizes -->
-      <app-drawer id="drawer" swipe-open>
+      <app-drawer id="drawer" swipe-open="[[smallScreen]]">
         <template is="dom-if" if="[[smallScreen]]">
           <app-toolbar class="profileBar">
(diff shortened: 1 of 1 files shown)


Lange:
I signed it!

keanulee:
LGTM. Thanks @Lange

PR #457, opened 2017-04-18
Fouc issues

keanulee:


frankiefu:
- `app-header-layout/demo/footer.html` demo doesn't seem to work. The footer should be at the bottom. I think we may need to set `@apply --layout-vertical; @apply --layout-fit;` on `div#wrapper` when `fullbleed` is set on `app-header-layout`.
- This is not related to this PR but can you also fix the `app-header-layout/demo/music.html` demo?  Seems like dom-repeat is not generating the song items.  Maybe we need to put the whole demo in a `dom-module`.

frankiefu:
I can see fouc when running `app-drawer-layout/demo/index.html` against Polymer 1.x.  Maybe we should replace the [raf debouncer in resetLayout for Polymer 1.x](https://github.com/PolymerElements/app-layout/blob/5e166f5b64f8e39037814a140a75374355aa262d/app-layout-behavior/app-layout-behavior.html#L63) wth `this.debounce` which is what [app-drawer-layout](https://github.com/PolymerElements/app-layout/blob/master/app-drawer-layout/app-drawer-layout.html#L226) and [app-header-layout](https://github.com/PolymerElements/app-layout/blob/master/app-header-layout/app-header-layout.html#L171) use in app-layout v1.

app-header-layout/app-header-layout.html
```diff
@@ -168,10 +186,12 @@
         var headerHeight = header.offsetHeight;
         // Update the header position.
         if (!this.hasScrollingRegion) {
-          var rect = this.getBoundingClientRect();
-          var rightOffset = document.documentElement.clientWidth - rect.right;
-          header.style.left = rect.left + 'px';
-          header.style.right = rightOffset + 'px';
+          requestAnimationFrame(function() {
```
keanulee: I've been using demo/demo2.html with Polymer 2. Note that this only happens when there is exactly 1 resize event between wide and narrow (e.g. rotating a phone, or toggling the dev tools pane).

PR #458, opened 2017-04-19
Pattern for Material Design Responsive Toolbar

frankiefu:
- will add the live demo to https://polymerelements.github.io/app-layout/patterns/md-responsive-toolbar later

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,10 @@ As more screen space is available, side navigation can transform into tabs.
 Content cards may expand to take up more horizontal space.
 ([Demo](https://polymerelements.github.io/app-layout/patterns/expand-card/index.html) - [Source](/patterns/expand-card/index.html))
 
+- **Material Design Responsive Toolbar:**
+Toolbar changes its height and padding to adapt mobile screen size.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -141,6 +141,10 @@ As more screen space is available, side navigation can transform into tabs.
 Content cards may expand to take up more horizontal space.
 ([Demo](https://polymerelements.github.io/app-layout/patterns/expand-card/index.html) - [Source](/patterns/expand-card/index.html))
 
+- **Material Design Responsive Toolbar:**
+Toolbar changes its height and padding to adapt mobile screen size.
+([Demo](https://polymerelements.github.io/app-layout/patterns/md-responsive-toolbar/index.html) - [Source](/patterns/md-responsive-toolbar/index.html))
```
keanulee: Are we still deploying to GH pages? Maybe should just link to the source?

frankiefu: We need a place to host all the live demos for patterns/templates, right?

keanulee: Wonder if there's a way we can get demos/ and patterns/ to show up on webcomponents.org (templates/ are a little different, since they're set up as self-contained applications).

PR #466, opened 2017-04-26
Tie drawer toggle to tap event

### LINKED ISSUE #465: Drawer toggle element responds to click event, not tap
When using the app-drawer-layout in a progressive application, the handler that toggles the drawer listens to the `click` event, not the `tap` event: [code here](https://github.com/PolymerElements/app-layout/blob/222ae2cc99cc8111ddc188b05791b80edc55cb0d/app-drawer-layout/app-drawer-layout.html#L211)

If the listener is changed to `tap`, the original behavior is preserved, and the drawer opens more quickly on mobile devices since the `tap` event bypasses the 300ms click delay. Since toggling the drawer is a common operation on mobile, it makes sense for this behavior to tie in to the `tap` event.

tmlbl:
Closes #465

### DIFF
--- a/app-drawer-layout/app-drawer-layout.html
+++ b/app-drawer-layout/app-drawer-layout.html
@@ -208,7 +208,7 @@ Custom property                          | Description                          
       },
 
       listeners: {
-        'click': '_clickHandler'
+        'tap': '_clickHandler'
       },
 
       observers: [


tmlbl:
I signed it!

keanulee:
Closing - see https://github.com/PolymerElements/app-layout/issues/465#issuecomment-298401945

PR #471, opened 2017-05-10
Make app-drawer-layout use --app-drawer-width to remove measurement

keanulee:


frankiefu:
Running `demo/demo3.html`, open/close the drawer doesn't transition.

PR #494, opened 2017-06-26
Update to point to actual demo

samuelli:


jsilvermist:
I see that the comment has a space between it and the dom-module, will this be a problem after Polymer/polymer-analyzer#696, or does that not apply to this case as it's the main comment at the top? (Not including @license)

samuelli:
@aomarks confirmed it won't be an issue, since that PR only concerns JS comments, not HTML comments.

PR #495, opened 2017-07-05
Avoid style recalc by reordering operations

keanulee:


frankiefu:
While testing the PR I found an issue in `app-drawer` such that, when it's placed inside `app-drawer-layout`, doesn't transition to open/close state. This only happens if you start the app in narrow mode. You can see the problem by launching `demo/demo2.html` in mobile screen size and then tap on the menu button to open the drawer.

PR #501, opened 2017-08-15
Declare AppLayout._scrollTimer early and clearly.

rictic:
One more tooling-related change!


rictic:
Alternatively, we could make it a local variable, as it seems like it's only used in this file.

PR #516, opened 2017-10-10
Fix annotations

valdrinkoshi:
Makes closure compiler happy 🎉


valdrinkoshi:
Can we have this merged & released?

PR #520, opened 2017-10-26
'<' and '>' -> '&lt;' and '&gt;'

davidmaxwaterman:
Elements were being interpreted as html and not quoted, making the text nonsense.
I *think* this should fix that, but I can't think how to test it.

ref:
https://www.webcomponents.org/element/PolymerElements/app-layout/elements/app-drawer
```
NOTE:* If you use with and specify a value for --app-drawer-width, that value must be accessible by both
elements. This can be done by defining the value on the :host that contains (or html if outside a shadow
root):
```

### DIFF
--- a/app-drawer/app-drawer.html
+++ b/app-drawer/app-drawer.html
@@ -45,9 +45,9 @@ Custom property                  | Description                            | Defa
 `--app-drawer-content-container` | Mixin for the drawer content container | {}
 `--app-drawer-scrim-background`  | Background for the scrim               | rgba(0, 0, 0, 0.5)
 
-**NOTE:** If you use <app-drawer> with <app-drawer-layout> and specify a value for
+**NOTE:** If you use `<app-drawer>` with `<app-drawer-layout>` and specify a value for
 `--app-drawer-width`, that value must be accessible by both elements. This can be done by
-defining the value on the `:host` that contains <app-drawer-layout> (or `html` if outside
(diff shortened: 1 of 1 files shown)


keanulee:
That does not fix it (you can see how it's rendered in GitHub, like https://github.com/davidmaxwaterman/app-layout/blob/patch-1/app-drawer/README.md). Wrapping it in backticks should work, like `<app-drawer>`.

davidmaxwaterman:
Thanks @keanulee. I was pleased to learn of the way to preview the file :)

I tried to use back ticks instead, even though I wasn't entirely convinced they would give the correct effect - ie it seemed like the original author just wanted it in plain text, and using back ticks adds additional style. In any case, my attempts seem to have failed. It looks like I've done "the same thing" as on the "--app-drawer-width" text (to name just one), but the preview link you gave still doesn't render the text.
Actually, it looks like the preview link you gave me is for a different file - it's the README.md file. Surely that is built somehow from the html file?

Render me confused, and ready to be instructed :)

keanulee:
Fixed in https://github.com/PolymerElements/app-layout/commit/4c621c45bc8d7d47563e9e3fb5ac3120b21426c2. The backticks fix it because it forces the markdown renderer into "code mode" which doesn't interpret `<` as the start of an HTMLElement.

The docs are (manually) replicated in README and source, and webcomponents.org uses the README, so generally both should be updated.

The original author (me) didn't really pay attention to whether backticks were required 😛

PR #527, opened 2017-11-21
create app-layout.js

itspiratefairy:
'app-layout.js' module which can be imported by an other module to include all elements under 'app-layout' group

### DIFF
--- a/app-layout.js
+++ b/app-layout.js
@@ -1,0 +1,18 @@
+/*
+@license
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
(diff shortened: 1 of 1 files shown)


itspiratefairy:
i signed it

TimvdLippe:
This file is auto-generated by the modulizer to https://github.com/PolymerElements/app-layout/blob/__auto_generated_3.0_preview/app-layout.html If you would like to request changes on how the Polymer 3 version of this element is generated, feel free to file a bug at https://github.com/polymer/polymer-modulizer

PR #540, opened 2018-03-19
Add TypeScript declarations to app-layout

TimvdLippe:
This PR adds TypeScript declarations generated by https://github.com/Polymer/gen-typescript-declarations/

These declarations can be re-generated by running `npm run update-types`.

Tracker: https://github.com/Polymer/gen-typescript-declarations/issues/79


TimvdLippe:
I can't seem to get CI passing. Initially I ran the `Travis` pass from Tedium, but that generated a lot of changes which seemed to make Travis unhappy. Then I took the `paper-button` configuration (https://github.com/PolymerElements/paper-button/blob/master/.travis.yml), but even that keeps on failing. It seems to be the `Chrome headless` issue again.

TimvdLippe:
Hm, it appears that I was watching the incorrect Travis build. Latest commit seems to be working now

aomarks:
I think we want to exclude all the demos and templates, so you'd need something like this in `gen-tsd.json`:

```json
"excludeFiles": [
  "index.html",
  "**/test/**",
  "**/demo/**"
]
```

TimvdLippe:
@aomarks Thanks, done!

PR #545, opened 2018-04-10
Automatic format!

e111077:
Runs the experimental autoformatter [webmat](https://github.com/PolymerLabs/webmat) which runs clang-format on js files and makes it so that it can run on HTML files. Please look through this PR with `?w=1` in the diff to make sure that no logic was changed.

What has changed?
You can run the formatter on the whole project by running `npm run format` and ex/include files from the formatter, and enter your custom clang-format config in a formatconfig.json see webmat readme for more


app-box/app-box.html
```diff
@@ -135,14 +135,9 @@ <h2>Sub title</h2>
     Polymer({
       is: 'app-box',
 
-      behaviors: [
-        Polymer.AppScrollEffectsBehavior,
-        Polymer.IronResizableBehavior
-      ],
+      behaviors: [Polymer.AppScrollEffectsBehavior, Polymer.IronResizableBehavior],
```
keanulee: At what point does the formatter decide to have each item on its own line (`behaviors`, `listeners`, and `properties`)? This seems like it should be.

e111077: It's using google styleguide, so I think it's after it hits 80 cols

app-drawer/app-drawer.d.ts
```diff
@@ -11,153 +11,3 @@
...
-   */
-  resetLayout(): void;
-  _setKeyboardFocusTrap(): void;
-  _tabKeydownHandler(event: any): void;
-  _openedPersistentChanged(opened: any, persistent: any): void;
-}
-
-interface HTMLElementTagNameMap {
-  "app-drawer": AppDrawerElement;
-}
```
keanulee: The typings for app-drawer are now missing

app-drawer/app-drawer.html
```diff
@@ -283,8 +255,9 @@
           this._boundEscKeydownHandler = this._escKeydownHandler.bind(this);
           this.addEventListener('keydown', this._tabKeydownHandler.bind(this))
 
-          // Only listen for horizontal track so you can vertically scroll inside the drawer.
-          this.listen(this, 'track', '_track');
+              // Only listen for horizontal track so you can vertically scroll
```
keanulee: Missing semicolon above is causing this indentation.

app-drawer/app-drawer.html
```diff
@@ -502,8 +475,8 @@
 
         // Calculate the amount of time needed to finish the transition based on the
         // initial slope of the timing function.
-        var t = this._FLING_INITIAL_SLOPE * dx / velocity
-        this._styleTransitionDuration(t);
+        var t = this._FLING_INITIAL_SLOPE * dx /
+            velocity this._styleTransitionDuration(t);
```
keanulee: Missing semicolon

app-header/app-header.html
```diff
@@ -314,16 +314,14 @@ <h1 main-title>App name</h1>
...
+      behaviors: [Polymer.AppScrollEffectsBehavior, Polymer.AppLayoutBehavior],
 
       properties: {
         /**
          * If true, the header will automatically collapse when scrolling down.
-         * That is, the `sticky` element remains visible when the header is fully condensed
-         * whereas the rest of the elements will collapse below `sticky` element.
+         * That is, the `sticky` element remains visible when the header is fully
+         *condensed whereas the rest of the elements will collapse below `sticky`
+         *element.
```
keanulee: nit: no space between `*` and text. Not a blocker, but curious why linter doesn't handle this.

app-scroll-effects/app-scroll-effects-behavior.html
```diff
@@ -14,22 +14,25 @@
...
    *
-   * app-layout provides a set of scroll effects that can be used by explicitly importing
-   * `app-scroll-effects.html`:
+   * app-layout provides a set of scroll effects that can be used by explicitly
+   * importing `app-scroll-effects.html`:
    *
    * ```html
-   * <link rel="import" href="/bower_components/app-layout/app-scroll-effects/app-scroll-effects.html">
+   * <link rel="import"
+   * href="/bower_components/app-layout/app-scroll-effects/app-scroll-effects.html">
```
keanulee: This _might_ be an issue with the README generator (i.e. will it remove the `* ` in the HTML block), but maybe it's ok?

app-drawer/app-drawer.d.ts
e111077: that. is. not. supposed. to. happen. 😕

app-header/app-header.html
e111077: odd.. it should

app-drawer/app-drawer.d.ts
e111077: ah this is because the missing semicolons caused a typescript syntax error

app-scroll-effects/app-scroll-effects-behavior.html
e111077: I can tell it to not reflow comments

e111077: Sheesh, nevermind, that's a bad bad flag
image

keanulee: I think it's better if it enforces the line limits on all comments - code snippets in comments are a small use case, and I'm not even sure if the README generator is used on this repo anyways. Suggest you ignore this issue for now, and we'll deal with it if it's an issue later.

e111077:
I was testing you

keanulee:
LGTM

PR #560, opened 2018-06-05
Automatic cleanup!

e111077:


e111077:
@keanulee this should be fixed now. It seems like the failures are just now FF test failing (and it displaying ungracefully in WCT)

PR #563, opened 2018-06-29
3.0 preview dev

keanulee:


package.json
```diff
@@ -8,12 +8,8 @@
   "name": "@polymer/app-layout",
   "license": "BSD-3-Clause",
   "devDependencies": {
-    "@polymer/gen-typescript-declarations": "^1.2.2",
-    "bower": "^1.8.0",
-    "webmat": "^0.2.0",
-    "es6-promise": "^4.0.0",
     "@polymer/font-roboto": "^3.0.0-pre.20",
-    "@polymer/iron-component-page": "^3.0.0-pre.20",
+    "@polymer/iron-ajax": "^3.0.0-pre.20",
```
e111077: Why did this get added?

keanulee: Some of the demos use it (e.g. https://github.com/PolymerElements/app-layout/blob/__auto_generated_3.0_preview/app-grid/demo/flickr-grid-layout.html). Honestly don't know why it wasn't there before - it should be.

Separately, I removed `es6-promise` because wcjs v2 has Promise polyfill. (Previously it had to support hybrid mode and neither wcjs v0 nor iron-request v2 had the polyfill.)

PR #579, opened 2019-07-01
add padding var

43081j:
what do you think of this? we have a problem at the min where we can't pass a mixin in because `@apply` isn't supported on our end. so we can't get rid of/change the default padding of the drawer.

would this work? let me know what you think

cc @aomarks


43081j:
Thanks for taking a look. 

Will add the docs in the morning 👍

43081j:
here you go, let me know if you want me to add anything else

e111077:
release awaiting internal testing

e111077:
published under 3.1.0

